PR #12, opened 2015-04-03
Update min(), max(), abs(), ADD stream.adoc

sh-aks:
Update min(), max(), abs() acc to https://github.com/arduino/Arduino/issues/2602, 
ADD stream.adoc missing before.


sh-aks:
@agdl Updates for min(), max(), abs() and Added stream.adoc

PR #23, opened 2015-12-10
Fix `int` documentation on overflow (which is UB)

cousteaulecommandant:
The documentation says that if you add 1 to an int variable whose value is 32767 it'll just "roll over" to -32768, but this is actually undefined behavior (for signed integers of any length) and can't be trusted in behaving consistently.  Example:

```
boolean will_overflow(int x) {
    int y = x+1;
    return y < x; // compiler assumes this will never happen => will always return false
}

    will_overflow(INT_MAX)  // returns false instead of true
```

Notice that this is not the case for unsigned integer types, which are required to "roll over" as documented.

### DIFF
--- a/Language/Variables/Data Types/int.adoc
+++ b/Language/Variables/Data Types/int.adoc
@@ -53,17 +53,8 @@ The Arduino takes care of dealing with negative numbers for you, so that arithme
 
 [float]
 === Notes and Warnings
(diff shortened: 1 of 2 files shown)


cousteaulecommandant:
Alternatively, reject this PR and instead fix Arduino so that it adds the `-fwrapv` flag, which makes signed integers behave in the currently documented way (which is probably easier to understand for inexperienced programmers so it might be a better option).

PR #44, opened 2017-07-13
Update equalTo.adoc

SimonePDA:
Corrected the end of page links


smellai:
I fixed a link, but why 'equal to' is related to if, char, int and long? @SimonePDA

PR #45, opened 2017-07-14
Update greaterThan.adoc

SimonePDA:
Corrected the see also links for the whole set of operators


smellai:
Same here, I think equalTo, notEqualTo and lessThan was better

PR #76, opened 2017-10-10
Update bitshiftRight.adoc

kengdahl:


Language/Structure/Bitwise Operators/bitshiftRight.adoc
```diff
@@ -93,3 +93,20 @@ int y = x >> 3;   // integer division of 1000 by 8, causing y = 125.
```
smellai: remove the see also section inside how to use

PR #104, opened 2017-10-12
Update else.adoc

kengdahl:
moved see also to section of its own.


Language/Structure/Control Structure/else.adoc
```diff
@@ -70,10 +70,18 @@ else
...
 
+--
+// HOW TO USE SECTION ENDS
+
+
+
+
+// SEE ALSO SECTION BEGINS
+[#see_also]
+
```
smellai: missing dashes

PR #126, opened 2017-10-12
Update byteCast.adoc

kengdahl:
Moved see also to section of its own.


Language/Variables/Conversion/byteCast.adoc
```diff
@@ -56,3 +56,21 @@ Converts a value to the link:../../data-types/byte[byte] data type.
```
smellai: please remove see also section inside how to use

PR #139, opened 2017-10-30
Update float.adoc

SimonePDA:
from fpConstants to fpconstants


smellai:
the correct link is `floatingpointconstants` as we said yesterday

PR #271, opened 2017-11-25
Fixed page title, but the file needs revision

robsoncouto:
May need changes:
-I do not know if the " +" in first paragraph is really supposed to be a newline or a hard-break.
-if(serial) in the list of functions is in the top, out or alphabetical order. It was probably added later in 1.0.1, but I don't know if it is there for emphasis.
-The * in the example list appear in the reference webpage while not visible in many other pages

### DIFF
--- a/Language/Functions/Communication/serial.adoc
+++ b/Language/Functions/Communication/serial.adoc
@@ -1,14 +1,10 @@
 ---
-title: serial
+title: Serial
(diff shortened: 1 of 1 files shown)


robsoncouto:
Ops. Just noticed that "Serial()" is not rendered in the reference, page the value of the title tag (serial). It would be better a Serial tough I guess, as it is a class.

PR #276, opened 2017-11-30
Remove First charater. May be a BOM?

sebuls:


robsoncouto:
Fixed in ab9210139534791e9a06a4de7f4c62941ad5d428.

PR #303, opened 2018-01-21
Append volatile.adoc with info about atomic read

hydrocontrol:
I think the explanation of volatile needs to include some warning about the use of volatile ints and longs. 
I realize that this is an advanced topic, but hey, the use of "volatile" is advanced, so a heads up can't hurt :-)
Thank you for your consideration.
Thomas

### DIFF
--- a/Language/Variables/Variable Scope & Qualifiers/volatile.adoc
+++ b/Language/Variables/Variable Scope & Qualifiers/volatile.adoc
@@ -85,10 +85,11 @@ void blink()
 [source,arduino]
 ----
 #include <util/atomic.h> // this library includes the ATOMIC_BLOCK macro.
-
+volatile int input_from_interrupt
 
   ATOMIC_BLOCK(ATOMIC_RESTORESTATE) {
      // code with interrupts blocked (consecutive atomic operations will not get interrupted)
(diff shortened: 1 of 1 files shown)


hydrocontrol:
Dear SimonePDA,

 

I‘m not sure if I understand the function of Github correctly - you have merged the changes which I proposed, but they are not showing up on https://www.arduino.cc/reference/en/language/variables/variable-scope--qualifiers/volatile/ ?

Is this simply a delay, or is there a problem?

 

I am asking because my other proposal appears fine in the arduino reference.

 

Thank you and kind regards

Thomas

 

 

http://www.hydroelectronics.net
electronic load controllers and induction generator controllers

 
 

Gesendet: Montag, 22. Januar 2018 um 09:22 Uhr
Von: SimonePDA <notifications@github.com>
An: arduino/reference-en <reference-en@noreply.github.com>
Cc: hydrocontrol <th.schuler@gmx.de>, Author <author@noreply.github.com>
Betreff: Re: [arduino/reference-en] Append volatile.adoc with info about atomic read (#303)


Merged #303.

hydrocontrol:
It's online now. Seems I was just too impatient :-)

PR #304, opened 2018-01-21
Update unsignedInt.adoc

hydrocontrol:
I'm unsure how deep the language reference is supposed to dig into the inner workings of the MCU. 
I think in this case the issue with scope of the calculation is relevant, because it is a consequence of the rollover. For beginners, the easiest is to avoid unsigned variables alltogether :-)
Thank you for considering
Thomas


SimonePDA:
A useful explanation about the math behind signed and unsigned variables

Issue #317, opened 2018-03-09
Output from example code on Serial.print() is messy


shane-lawson:
The sample code at the page https://github.com/arduino/reference-en/blob/master/Language/Functions/Communication/Serial/print.adoc yields 
![image](https://user-images.githubusercontent.com/25392583/37228299-0c8e15ec-23d8-11e8-9c13-ca1c26bbb034.png) 
in the Arduino serial monitor. This is messy and difficult to tell what's going on if you're not familiar with the serial output.

SimonePDA:
Fixed with pull request  #345

PR #318, opened 2018-03-09
Update sample code to give expected output

shane-lawson:
Solves issue #317.

The change in line 97 is because the 0th iteration of the `for` loop 
would print on the same line as the column headers, the newline 
vs. tab means this starts on line below the headers.

The change in line 103 is because the header "NO FORMAT" 
extends past the first indent position, causing the outputs to 
line up immediately under incorrect columns headers, and only 
with much care does one avoid the mistake of reading the 
wrong column based on the header.

The change in line 22 is to set the text as code in the display.


SimonePDA:
Verified the update and it prints well.

Thank you!

PR #320, opened 2018-03-21
Fix Serial.available() example

diamondo25:


SimonePDA:
The current Mega example is working and clearly shows the input and output process. It con be shortened as suggested, but it is less useful for the learning user. Moreover as it is not broken, the term "Fix" is not the proper one. It is a suggestion.

diamondo25:
Nope, the code does not work in the current version of Arduino, as the BYTE keyword is not supported anymore.

SimonePDA:
As for the first example,  we will change BYTE to DEC, fixing the code without altering the basic purpose of showing an input and an output on two different serial streams.

Thanks for the "BYTE" fix that wasn't clearly pointed out as the reason of the fix. 
We have to go through many pull requests and sometimes without any comment given, it is very difficult for us to understand which is the reason why a pull request is made.

PR #344, opened 2018-03-28
Update available.adoc

SimonePDA:
Fixes the issue with BYTE keyword that is no longer supported, fixing #320


diamondo25:
However, it will now write any value as ascii number back, not as-is as the example would think you it does. Therefor using .write() is still a better solution than .print()

PR #350, opened 2018-04-18
Rename "modulo" to "remainder"

cousteaulecommandant:
When negative numbers are involved, *"modulo"* typically refers to an operation that has the same sign as the second operand, whereas *"remainder"* has the sign of the first (this distinction is clear e.g. in the Ada language, which has both `mod` and `rem` operators).  Specifically, the standard refers to the result of `%` as "remainder", not "modulo".
In Python, for example, `(-11) % 10` is 9, but in C and C++ it is -1).  This can be a problem if `%` is being used to restrict an array index to a range (e.g. to implement a circular buffer), if said index can be negative.

I have replaced all occurrences of "modulo" with "remainder" in `modulo.adoc`, and provided two examples of what happens when you use `%` with negative operands.  However I have left the file name as "`modulo.adoc`" (because honestly I have no idea what happens if you mess with the file names).

### DIFF
--- a/Language/Structure/Arithmetic Operators/modulo.adoc
+++ b/Language/Structure/Arithmetic Operators/modulo.adoc
@@ -1,6 +1,6 @@
 ---
 title: "%"
-title_expanded: "modulo"
+title_expanded: "remainder"
 categories: [ "Structure" ]
 subCategories: [ "Arithmetic Operators" ]
 ---
@@ -9,7 +9,7 @@ subCategories: [ "Arithmetic Operators" ]
 
 
 
-= % Modulo
+= % Remainder
 
 
 // OVERVIEW SECTION STARTS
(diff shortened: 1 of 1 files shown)


SimonePDA:
I have to ask to the team if they agree on this change.

Artoria2e5:
An argument for this terminology can be that it is consistent with C/C++ standard (http://port70.net/~nsz/c) descriptions, but conversely it should be noted that C++11 calls it "modulus" too. Nevertheless "remainder" wins out at accuracy and familiarity.

cousteaulecommandant:
Should the file be renamed to `remainder.adoc`?  Now it's everything fine except that the top of the documentation still says *"Reference > Language > Structure > Arithmetic operators > **Modulo**"* (I don't know if just changing the file name fixes this; I assumed the `title_expanded:` field would take care of this.)

SimonePDA:
I took care of that change. Thank you for pointing that out.

cousteaulecommandant:
It would seem that `remainder.adoc` was successfully created, but the commit that removes `modulo.adoc` failed to build; as a result, there are now two entries for "% (remainder)" in https://www.arduino.cc/reference/en/

SimonePDA:
Yes, I know. It is a bug about deletions. We will take care of it after this holidays.

PR #357, opened 2018-04-18
Link to avr-libc documentation for PROGMEM funcs

Artoria2e5:
This change adds a link to the avr-libc documentation of progmem functions, so that folks can see all kinds of functions available and choose the one that fits them most. This also answers where all the functions used here come from.


Artoria2e5:
Oh there actually is a broken link to that page somewhere in the doc. Closing PR to send an actual syntax fix.

PR #363, opened 2018-05-13
abs() functions should always return a positive value

amitabhkant:


robsoncouto:
Hi @amitabhkant , if x is less than zero, abs() should return the negative of that. As the negative of a negative value, which is positive.

amitabhkant:
Oh.. My bad ... I was not reading the documentation correctly then..

robsoncouto:
No problem at all. Thanks for your interest in contributing to the Arduino reference!

PR #383, opened 2018-06-26
Added Notes & Warnings for map() function

theintel:
Included potential issues that may be encountered due to the use of map() function. As map() always returns integer, significant error might appear during value conversions.


theintel:
[This issue](https://github.com/arduino/Arduino/issues/2466) from 2014 contains an in-depth discussion on the precision issue of the map() function.

PR #392, opened 2018-07-19
Fixed strange BOM character in isAscii.adoc

SeppPenner:


SeppPenner:
Ok, this is done via https://github.com/arduino/reference-en/pull/449

Issue #393, opened 2018-07-19
Strange BOM character in isControl.adoc file


SeppPenner:
![image](https://user-images.githubusercontent.com/9639361/42928170-e0d1e85e-8b36-11e8-9640-cf23c53d8728.png)
And another one.

SeppPenner:
Fixed by https://github.com/arduino/reference-en/pull/394

SeppPenner:
Already fixed via https://github.com/arduino/reference-en/pull/449

PR #394, opened 2018-07-19
Fixing Strange BOM character in isControl.adoc

SeppPenner:


SeppPenner:
Fixes https://github.com/arduino/reference-en/issues/393

SeppPenner:
Ok, this is done via https://github.com/arduino/reference-en/pull/449

Issue #434, opened 2018-09-13
const.adoc does not reference to #define properly


SeppPenner:
https://github.com/arduino/reference-en/edit/master/Language/Variables/Variable%20Scope%20&%20Qualifiers/const.adoc

#define is only referenced correctly once.

robsoncouto:
#434 was not accepted because there were enough mentions to the #define page already

SeppPenner:
@robsoncouto In my opinion it would make sense to reference all occurences of these. But I will open a discussion in the issues.

SeppPenner:
I have created an issue for a general discussion on this. In my opinion it would make sense to link every occurence of functions or constants. https://github.com/arduino/reference-en/issues/462

PR #435, opened 2018-09-13
Fixed reference to #define.

SeppPenner:
Fixes https://github.com/arduino/reference-en/issues/434


robsoncouto:
Hi @SeppPenner, just a reference to the #define page in the text is enough. 
Besides, there is another reference in the See also section.

SeppPenner:
I have created an issue for a general discussion on this. In my opinion it would make sense to link every occurence of functions or constants.

Issue #441, opened 2018-09-19
doWhile.adoc does not increment counter variable in example code


SeppPenner:
https://github.com/arduino/reference-en/blob/master/Language/Structure/Control%20Structure/doWhile.adoc

![grafik](https://user-images.githubusercontent.com/9639361/45754615-625a6e80-bc1c-11e8-97a6-ce232404eefa.png)

The example code simply doesn't make sense without increments because the loop will run infinitely...

SeppPenner:
As @Artoria2e5 stated (https://github.com/arduino/reference-en/pull/442), this is bullshit and I was confused. Forget this one.

PR #442, opened 2018-09-19
Fixing doWhile.adoc does not increment counter

SeppPenner:
Fixes https://github.com/arduino/reference-en/issues/441


Artoria2e5:
`x = readSensors()` makes this pointless.

SeppPenner:
You're right with that one, but there is another error in the code. I will add another PR and issue on this.

Issue #443, opened 2018-09-19
break.adoc example code doesn't make sense without a threshold


SeppPenner:
https://github.com/arduino/reference-en/edit/master/Language/Structure/Control%20Structure/break.adoc

The example code does not set a variable `threshold`. I suggest to use a fixed value of e.g. 40.

PR #444, opened 2018-09-19
Fixed break.adoc example code doesn't make sense

SeppPenner:
Fixes https://github.com/arduino/reference-en/issues/443


robsoncouto:
@SeppPenner , the code in this page is just a snippet. Not a fully functioning sketch, threshold is clearly a variable previously defined. 

The Example Code section also states:

`
In the following code, the control exits the for loop when the sensor value exceeds the threshold.
`

I do understand it  may seem confusing to a beginner. So I added the variable declaration on the top of the code :)

SeppPenner:
@robsoncouto That was my intention. I understand the code as well as it was before. But as you said, it might be confusing for beginners.

PR #459, opened 2018-10-04
Missing a break in example

fv3rdugo:
In the example, the default option is empty, it has a comment line, this would generate an error:
> expected primary-expression before the '}' token

Normally, the default option has statements, but if it is not the case, insert a break line is a solution
```
switch (estat) {
    case 'A':
      // statements
      break;
    case 'M':
      // statements
      break;
      default:
      // statements
      break;  //not necessary if you have statements
  }
```

### DIFF
--- a/Language/Structure/Control Structure/switchCase.adoc
+++ b/Language/Structure/Control Structure/switchCase.adoc
@@ -37,6 +37,7 @@ switch (var) {
     break;
   default:
     // statements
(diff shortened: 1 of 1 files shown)


SeppPenner:
@fv3rdugo This is not necessary. The default case is always the last case in the switch case, so the switch-case-statement can never "fall through" this one...

PR #463, opened 2018-10-08
Updated analogRead() to use An style pin names

robsoncouto:
Also added a table with operating voltages, pin manes and resolution for common Arduino boards. Unfortunately I could not find the sampling rate for newer boards to add that in too. 
Related issue: #447


robsoncouto:
@per1234, could you please have a look at this?

PR #471, opened 2018-12-20
Got rid of the quote characters in visible on the Arduino website

lemio:
On the Arduino reference guide, there are quotes around the function Syntaxis, this might confuse the user, especially since it is not consistent, and present when copy pasting it.

image

I only checked the characters functions for now


robsoncouto:
These characters should not be there indeed. 
Thanks for spotting these out.

PR #626, opened 2019-06-16
SerialPrintConc.adoc

Sandromesquita:
Most Arduino IDE users use various Serial.print() functions to print on the Serial monitor screen a one line sentence, using Multiple Serial.print (). In this project present a way to get the same result but only writing one line of theSerial.print () function.

For you to concatenate multiple strings and variables on the same print line you simply use the '+' operator between the strings and if it is not a string, you must convert the variable you want to insert into a string. For example-

* `Serial.println(“Hello, today is day” + String(x) + “, good Morning!”)`

[float]
=== Syntax
`_Serial_.print(“String” + String(var))`

https://create.arduino.cc/projecthub/SandroMesquita/function-serial-print-in-only-one-line-29ef8d?ref=platform&ref_id=424_trending___&offset=6

### DIFF
--- a/Language/Functions/Communication/Serial/print.adoc
+++ b/Language/Functions/Communication/Serial/print.adoc
@@ -45,6 +45,14 @@ To send data without conversion to its representation as characters, use link:..
 `_Serial_.print(val)` +
 `_Serial_.print(val, format)`
 
+For you to concatenate multiple strings and variables on the same print line you simply use the '+' operator between the strings and if it is not a string, you must convert the variable you want to insert into a string. For example-
+
+* `Serial.println(“Hello, today is day” + String(x) + “, good Morning!”)`
+
+[float]
+=== Syntax
+`_Serial_.print(“String” + String(var))`
+
 
 [float]
 === Parameters


per1234:
I appreciate you taking the time to submit this pull request, but frivolous use of the wasteful and dangerous String class is not something we will promote in the Arduino Language Reference.

Sandromesquita:
Thanks for the reply, but what is the risk of concatenation?

per1234:
Sorry, I was working on a different project at the time of my last reply and didn't take the time to explain my reasoning well.

Here's a nice explanation of the dangers of using String:
https://majenko.co.uk/blog/evils-arduino-strings

It is possible to use String safely, but often the people who use String are the same people who are not experienced enough to understand how to use them safely,  what the dangers are, or how to troubleshoot the problems caused by memory fragmentation. In the case of just wanting to save yourself from writing a few extra lines of code, it's simply not worth it.

The example code from your Project Hub article compiles for Uno to 3312 bytes program memory and 222 bytes of SRAM for globals. The equivalent non-String code compiles to 1922 bytes program memory and 212 bytes of SRAM for globals. That's pretty significant overhead for very little benefit on one of the more limited microcontrollers that are so popular in the Arduino world. On one of the more resource rich microcontrollers, perhaps it's not so much of a concern.

We do already document String conversion and String concatenation in the String documentation and document that `Serial.print()` can take any data type. So, regardless of opinions on using String, I feel that it's not necessary to document this in the `Serial.print()` reference page since the information is already available in the existing documentation.

Please don't take this as a discouragement from making further contributions to the Arduino Language Reference. There is a lot of room for improvement and I very much welcome any help anyone wants to give.

Sandromesquita:
Ok. I understand. Thankyou for explication.

What's you Facebook for I looking your projects?

Sandromesquita:
Understand, very good.


Em seg, 17 de jun de 2019 às 06:01, per1234 <notifications@github.com>
escreveu:

PR #671, opened 2019-07-13
Code Highlight markup and Link colouration fix sizeof.adoc

animeshsrivastava24:
Line 69 Do we need to modify the **ints** to **int** and add a code element with link?
Ref #546

### DIFF
--- a/Language/Variables/Utilities/sizeof.adoc
+++ b/Language/Variables/Utilities/sizeof.adoc
@@ -23,12 +23,12 @@ The `sizeof` operator returns the number of bytes in a variable type, or the num
 
 [float]
 === Parameters
-`variable`: The thing to get the size of. Allowed data types: any variable type or array (e.g. `int`, `float`, `byte`).
+`variable`: The thing to get the size of. Allowed data types: any variable type or array (e.g. `link:../../data-types/int[int]`, `link:../../data-types/float[float]`, `link:../../data-types/byte[byte]`).
 
 
 [float]
 === Returns
-The number of bytes in a variable or bytes occupied in an array. Data type: `size_t`.
+The number of bytes in a variable or bytes occupied in an array. Data type: `link:../../data-types/size_t[size_t]`.
 
 --
 // OVERVIEW SECTION ENDS


per1234:
>Do we need to modify the ints to int and add a code element with link?

I would do it like this:
```
Note that `sizeof` returns the total number of bytes. So for arrays of larger variable types such as link:../../data-types/int[`int`]s, the for loop would look something like this.

**Comment by animeshsrivastava24** [comment]
Oh, I didn't think over it in this manner. Thanks a lot for the suggestion.
I'll send another PR with **this** (link:../../data-types/int[`int`]s) update in it. Please merge this one.

**Comment by animeshsrivastava24** [comment]
@per1234  While working on to send the next PR, I am stuck at a small problem and need your guidance

I was working on to modify the [unsigned int](https://github.com/arduino/reference-en/blob/master/Language/Variables/Data%20Types/unsignedInt.adoc) file

For the first occurrence of ints,
I tried the following `link:../int[int]`s , this produces a problem related to desired display output

This produced 
image

While for the further occurrences of ints

link:../int[`int`]s , works fine
 
image

Please guide how to solve such first occurences

**Comment by per1234** [comment]
I found a solution. It's a bit ugly, but it does work:
```
`link:../int[int]`{empty}s
```
`{empty}` is a predefined character replacement attribute in asciidoctor:
https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#attributes-and-substitutions

animeshsrivastava24:
@per1234 Thanks a lot for the valuable suggestion.

In recent PRs  #677 #679, I accidentally added code markup to text. I thought to have your views before proceeding further to do any such further modifications.
Please guide me, do the instances of unsigned int and ints need to be added under code markup in  
Line 15 
Line 19
https://github.com/arduino/reference-en/blob/master/Language/Variables/Data%20Types/unsignedInt.adoc

per1234:
@animeshsrivastava24 yes please. Any time there is code in the reference, it should be in code markup. I think it's especially important in this case where a code word is made plural, because the markup makes it clear that only part of the word is code, and so prevents confusion (what's an "ints"?)

PR #728, opened 2020-04-25
Update analogWrite.adoc

Machine4000:
The documentation fails to mention that the Due PMW resolution is 12bits, 0 - 4095.  Upon reading this you would assume its between 0 - 255, and most people won't be measuring the pins duty cycle and will find that they cannot reach above 6% duty with 255.

### DIFF
--- a/Language/Functions/Analog IO/analogWrite.adoc
+++ b/Language/Functions/Analog IO/analogWrite.adoc
@@ -51,6 +51,8 @@ The `analogWrite` function has nothing to do with the analog pins or the `analog
 `pin`: the Arduino pin to write to. Allowed data types: `int`. +
 `value`: the duty cycle: between 0 (always off) and 255 (always on). Allowed data types: `int`.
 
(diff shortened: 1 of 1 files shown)


Machine4000:
Yes that does sound good but I think you also should note that the default
for the due seems to be set to 12 bit mode not 8 bit.

per1234:
Default resolution for `analogWrite` is 8 bit:

https://github.com/arduino/ArduinoCore-sam/blob/0840aa277a7a6519147fcf041b0b3c5ae5dcbac1/cores/arduino/wiring_analog.c#L26

```c
static int _writeResolution = 8;
```

Running this sketch on my Due:

```cpp
void setup() {
  analogWrite(2, 127);
}
void loop() {}
```

Results in this signal on pin 2:

image

So I don't think the proposed change can be accepted. However, I still think it would be worth emphasizing the ability to adjust this resolution via `analogWriteResolution`, as we discussed earlier. So I would welcome any proposals along those lines.

PR #763, opened 2020-07-23
Update doWhile.adoc

nickbunyan:


nickbunyan:
Example as previously written does not demonstrate loop logic adequately

per1234:
Hi @nickbunyan. Thanks so much for taking the time to make this proposal! My feeling is that this is not a beneficial change because `x` is not intended to be a loop counter. `x` contains the return value from `readSensors()`, which might be any value the sensors output, so it's not necessary to then increment the sensor output.

You are welcome to retain your fork of this repository in order to provide your own version of the documentation. I hope you will continue to make proposals for any changes to the Arduinon language reference you think would be beneficial.

nickbunyan:
Happy to accept the response as a valid point but suggest that the need to 
increment and/or read value is noted to stop the example being an endless loop.
Thanks.
Sent from my OnePlus7 Android using
fumblefinger(TM) Please excuse typos...

PR #767, opened 2020-08-14
Update goto.adoc

jackaroo-DE:
It should be considered that - as opposed to functions - a "goto"-code can never return to its calling point. So instead of using the "goto"-control, it is advisable to "break" or "continue" the nested structures and define and call a function instead.


per1234 (commented):
I think the decision on whether this sort of change is desirable is contingent on https://github.com/arduino/reference-en/issues/623.

If the Arduino Language Reference is intended to document the Arduino core API and give a brief overview of the common parts of C++, such as `goto`, pointing users to use external C++ references for details on the aspects of the Arduino Language that are inherited from C++, then I would say that this proposed change is not beneficial.

On the other hand, if we are going to  taking the approach that the Arduino Language Reference is the sole source of documentation for the Arduino Language, then it would be beneficial to provide more comprehensive documentation of the parts inherited from C++, though I would like to find a way to separate the more detailed information to avoid making the Arduino Language Reference less contributor friendly.

So my opinion is this PR should be put on hold until there is an official decision on https://github.com/arduino/reference-en/issues/623.

PR #771, opened 2020-08-19
Misusage of `strlen_p()`

Jeremaiha-xmetix:
There is no actual reason to put a method that counts characters inside a `for-loop`, such that every iteration will reevaluate the length of `signMessage`.


Jeremaiha-xmetix:
Thank you for your merge @alranel

Issue #739, opened 2020-05-03
Misleading documentation of PROGMEM


edgar-bonet:
Here are some excerpts of the [documentation of `PROGMEM`][pgm] (emphasis mine):

> # Description
>
> Store data in flash (program) memory **instead of** SRAM. [...] It tells the compiler "put this information into flash memory", **instead of** into SRAM, where it would normally go. [...]

These statements imply that the usage of `PROGMEM` is a trade-off between flash usage and RAM usage: you save some RAM at the cost of extra flash consumption. _This is wrong_. Any constant for which the compiler cannot optimize out the storage (which includes every string literal) is normally stored **in flash memory**. At program startup, the C runtime creates **an extra copy** of the constant in RAM, which is the copy that the program will use further on.

The purpose of `PROGMEM` is _not_ to store the data in flash “_instead of_” RAM: that data will be in flash no matter what. Its purpose is to _avoid the extra copy in RAM_: PROGMEM data is retrieved from flash on demand, only when needed, instead of being copied _en masse_ to RAM at startup.

This may seem like nitpicking, but I think it is worth emphasizing that PROGMEM has _zero cost_ in terms of flash usage: it is useful no matter how much free flash you have.

The page also contains this statement:

> It requires special methods [...] to read the data from program memory **back into SRAM**, so we can do something useful with it.

This is, quite often, incorrect. The compiler is smart enough to allocate most local variables to internal CPU registers rather than RAM. The `pgm_read_*()` macros then read the data from flash _into the CPU_, where accessing a regular constant would instead read that data from RAM. Nothing is moving from flash to RAM. That is, unless you read a whole array at a time, or have so many locals that the compiler cannot fit them into the register file.

My point is that the sentence above implies that data needs to be in RAM in order to be useful (which is wrong), and that PROGMEM incurs the cost of a flash-to-RAM copy on every access (also wrong). Truth is, PROGMEM access is very slightly slower than RAM access: 3&nbsp;CPU cycles per byte instead of 2. But most times you will pay the cost of the flash access _instead of_ – not _in addition to_ – the cost of RAM access.

Note that there is a flash-to-RAM copy in `Serial.println(F("..."))`, as the string has to land in the serial TX buffer. This copy, however, replaces a RAM-to-RAM copy you would have without the `F()` macro. So there is no _extra_ copy induced by the use of `F()`.

[pgm]: https://www.arduino.cc/reference/en/language/variables/utilities/progmem/#_the_f_macro

[...]

PR #787, opened 2020-09-30
Update PROGMEM.adoc

MalcolmBoura:
The strings have to be in flash whether they are loaded at start up (the default) or when needed. Hence using embed or the F macro should have no impact on the flash use. There is a very small time penalty, about 1 clock cycle per character + a function call I expect, but I don't know enough about how the Atmega works to be certain.


edgar-bonet:
Note that this PR partially addresses issue #739.

edgar-bonet:
I updated issue #739, as it has been partially fixed by this pull request.

PR #788, opened 2020-10-03
Update increment.adoc

TADefx:
Corrected formatting of the Syntax section, caused by "+" operator that caused the "++" to be hidden and cut-off the lines

### DIFF
--- a/Language/Structure/Compound Operators/increment.adoc
+++ b/Language/Structure/Compound Operators/increment.adoc
@@ -24,11 +24,8 @@ Increments the value of a variable by 1.
 
 [float]
 === Syntax
-[source,arduino]
-----
-x++;  // increment x by one and returns the old value of x
-++x;  // increment x by one and returns the new value of x
-----
+`x{plus}{plus};  // increment x by one and returns the old value of x` +
+`{plus}{plus}x;  // increment x by one and returns the new value of x`
 
 
 [float]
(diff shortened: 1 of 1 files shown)


Language/Structure/Compound Operators/increment.adoc
```diff
@@ -24,8 +24,11 @@ Increments the value of a variable by 1.
 
 [float]
 === Syntax
-`x++;  // increment x by one and returns the old value of x` +
-`++x;  // increment x by one and returns the new value of x`
+[source,arduino]
+----
+x++;  // increment x by one and returns the old value of x
+++x;  // increment x by one and returns the new value of x
+----
```
per1234, line 31:
```suggestion
`x{plus}{plus};  // increment x by one and returns the old value of x` +
`{plus}{plus}x;  // increment x by one and returns the new value of x`
```
I think this is a better approach because it makes it more obvious (to me at least) why the markup needs to be this way.

With your appoach, it's not clear why it doesn't follow the established markup convention for the syntax sections, which could lead to some confusion for future editors or even the fix being undone.

per1234 (changes requested):
Nice catch @TADefx! This definitely needs to be fixed right away.

I've provided a suggestion for an alternative approach for the fix. Please let me know what you think.

TADefx:
Thanks for the quick reply! Love it! Good fix! I didn’t know how to work around the + issue...thanks!!! Tim

per1234 (approved):

per1234:
Thanks so much @TADefx!

PR #791, opened 2020-10-20
Include Nano 33 and Portenta

ubidefeo:
These boards are lacking in this piece of documentation, possibly elsewhere

### DIFF
--- a/Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
+++ b/Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
@@ -1,7 +1,7 @@
 ---
 title: analogReadResolution()
 categories: [ "Functions" ]
-subCategories: [ "Analog I/O" ]
+subCategories: [ "Zero, Due & MKR Family" ]
 ---
 
 
--- a/Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
+++ b/Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
@@ -1,7 +1,7 @@
 ---
 title: analogReadResolution()
 categories: [ "Functions" ]
-subCategories: [ "Zero, Due, MKR family, Nano 33 (BLE and IoT) and Portenta" ]
+subCategories: [ "Analog I/O" ]
 ---
 
 
--- a/Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
+++ b/Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
@@ -1,7 +1,7 @@
 ---
 title: analogReadResolution()
 categories: [ "Functions" ]
-subCategories: [ "Zero, Due & MKR Family" ]
+subCategories: [ "Zero, Due, MKR family, Nano 33 (BLE and IoT) and Portenta" ]
 ---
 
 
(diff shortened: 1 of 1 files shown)


Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
```diff
@@ -1,7 +1,7 @@
 ---
 title: analogReadResolution()
 categories: [ "Functions" ]
-subCategories: [ "Zero, Due & MKR Family" ]
+subCategories: [ "Zero, Due, MKR family, Nano 33 (BLE and IoT) and Portenta" ]
```
per1234, line 4: This would cause the page to no longer be published at https://www.arduino.cc/reference/en/:

---
VirtualBox_Ubuntu_20_10_2020_16_58_01

---

The reason is because no subcategory of this name has been defined. One could be defined, but I think it's not a sustainable strategy to continuously append board names as needed.

Better would be to change to a non-board specific category name. I would prefer to just move the `analogReadResolution()` and `analogWriteResolution()` reference pages to the "Analog I/O" subcategory, then make sure the reference pages are very clear about which boards support these functions.

What do you think?

Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
```diff
@@ -16,11 +16,12 @@ subCategories: [ "Zero, Due & MKR Family" ]
...
 [float]
 === Description
-analogReadResolution() is an extension of the Analog API for the Arduino Due, Zero and MKR Family.
+analogReadResolution() is an extension of the Analog API for the Zero, Due, MKR family, Nano 33 (BLE and IoT) and Portenta.
 
 Sets the size (in bits) of the value returned by `analogRead()`. It defaults to 10 bits (returns values between 0-1023) for backward compatibility with AVR based boards.
 
-The *Due, Zero and MKR Family* boards have 12-bit ADC capabilities that can be accessed by changing the resolution to 12. This will return values from `analogRead()` between 0 and 4095.
+The *Zero, Due, MKR family and Nano 33 (BLE and IoT)* boards have 12-bit ADC capabilities that can be accessed by changing the resolution to 12. This will return values from `analogRead()` between 0 and 4095.
+The Portenta H7 has a 16 bit ADC, which will allow values between 0 and 65535.
```
per1234, line 24:
```suggestion
The *Zero, Due, MKR family and Nano 33 (BLE and IoT)* boards have 12-bit ADC capabilities that can be accessed by changing the resolution to 12. This will return values from `analogRead()` between 0 and 4095. +
The *Portenta H7* has a 16 bit ADC, which will allow values between 0 and 65535.
```
- Line break to improve readability.
- Bold "Portenta H7" to match the formatting of the previous sentence.

Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
```diff
@@ -31,7 +32,7 @@ The *Due, Zero and MKR Family* boards have 12-bit ADC capabilities that can be a
 
 [float]
 === Parameters
-`bits`: determines the resolution (in bits) of the value returned by the `analogRead()` function. You can set this between 1 and 32. You can set resolutions higher than 12 but values returned by `analogRead()` will suffer approximation. See the note below for details.
+`bits`: determines the resolution (in bits) of the value returned by the `analogRead()` function. You can set this between 1 and 32. You can set resolutions higher than the supported 10,12 or 16 bits, but values returned by `analogRead()` will suffer approximation. See the note below for details.
```
per1234, line 35:
```suggestion
`bits`: determines the resolution (in bits) of the value returned by the `analogRead()` function. You can set this between 1 and 32. You can set resolutions higher than the supported 12 or 16 bits, but values returned by `analogRead()` will suffer approximation. See the note below for details.
```
As far as I can tell, all of the official boards with `analogReadResolution()` have either a 12 or 16 bit ADC. The core libraries of the 10 bit ADC AVR and megaAVR  boards don't have  `analogReadResolution()`.

per1234 (changes requested):

Language/Functions/Zero, Due, MKR Family/analogReadResolution.adoc
ubidefeo: makes a lot of sense, Per
I don't know how such content is organised, and I'm glad you posted your feedback 👍

ubidefeo: I have removed the category change and will post it as another PR

per1234 (approved):
Looks great. Thanks @ubidefeo!

PR #795, opened 2020-11-02
Unified links in tone.adoc.

SeppPenner:
https://github.com/arduino/reference-en/pull/792/files changed some links, but not all were added with `www.`


kengdahl:
@SeppPenner Thanks!

PR #798, opened 2020-11-14
Update if.adoc

kb7hunter:
The if condition (if true) will execute the FOLLOWING statement, not the preceding statement.

### DIFF
--- a/Language/Structure/Control Structure/if.adoc
+++ b/Language/Structure/Control Structure/if.adoc
@@ -16,7 +16,7 @@ subCategories: [ "Control Structure" ]
 --
 [float]
 === Description
-The `if` statement checks for a condition and executes the proceeding statement or set of statements if the condition is 'true'.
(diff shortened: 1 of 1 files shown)


kengdahl:
@per1234 I think this sounds better and is a simpler word. Could you see any reason why not to merge it?

per1234:
Even though the previous word **"proceeding"** was not necessarily wrong, I think this proposal is an improvement.

kb7hunter:
The word "preceeding" is definitely *wrong*, it refers to something that
happened BEFORE while the word "following" refers to something that is
AFTER the statement.
"Preceeding" is incorrect usage to describe an item that "follows" the
description. No ambiguity here, there is nothing preceding when the item
described follows the description.
Ken Hunter

per1234:
@kb7hunter you misread. The word is "proceeding", not "preceeding". Despite similar spellings, they have opposite meanings.

Regardless, I appreciate you taking the time to improve this documentation. Thanks!

PR #805, opened 2020-11-25
Update random.adoc

Sumanthdev:
The place of the print is mentioned in the document


Sumanthdev:
updated print location in ddescription.

PR #826, opened 2021-04-23
Update ifSerial.adoc

rohoog:
I got bitten by this delay in a loop that otherwise would easily keep up with interrupts. I think users should be aware of it! One would not expect such a line status return to take any time at all.


per1234 (changes requested):
Thanks for your suggestion @rohoog. Please move this information to the standard "**Notes and Warnings**" section at the standard position on the page as shown in the reference sample:

https://raw.githubusercontent.com/arduino/reference-en/master/AsciiDoc_sample/Reference_Terms/AsciiDoc_Template-Single_Entity.adoc

rohoog:
I moved the remark. Sorry for the delay.

PR #828, opened 2021-05-19
Add 'Signature' section to String.length() doc

brewmanz:
I just got failing unit tests, as it didn't like comparing integer literals and unsigned integers. 
Knowing where to find the definitions is useful, too. 
Other methods, and indeed all classes and functions, should adopt a similar philosophy.

### DIFF
--- a/Language/Variables/Data Types/String/Functions/length.adoc
+++ b/Language/Variables/Data Types/String/Functions/length.adoc
@@ -36,6 +36,10 @@ Returns the length of the String, in characters. (Note that this doesn't include
 === Returns
 The length of the String in characters.
 
+[float]
+=== Signature
+(in arduino/avr/cores/arduino/WString.h)
+inline unsigned int length(void) const;
 --
 // OVERVIEW SECTION ENDS
 


per1234:
Thanks for taking the time to submit a pull request @brewmanz .

The Arduino Language Reference has a standardized format for documenting function return and parameter types in a beginner friendly manner. Since this proposal diverges from that standardized format, we will be unable to accept it.

Regards, Per

brewmanz:
Hello Per

Thank you for that information; I guess I was looking in the wrong place.
I'm used to Language References being able to provide that kind of
information, that I was seeking to add.

Can you please advise me of the correct place to find the signature
information (or at least what the function returns) for various functions,
like String.length(), preferably inside the Arduino IDE ?

Regards

Brewmanz

per1234:
>I'm used to Language References being able to provide that kind of information, that I was seeking to add.

Adding the information is necessary and welcome. However, it's essential for it follow the established format to provide consistency across all the reference content.

>Can you please advise me of the correct place to find the signature
information

Unfortunately, the reference content is incomplete in regards to this important information. It is documented for some functions, but not others. When the Language Reference is lacking (or even incorrect in some cases), the best reference is the source code:
https://github.com/arduino/ArduinoCore-API/blob/1.2.0/api/String.h#L92

>preferably inside the Arduino IDE

The beta phase Arduino IDE 2.x does show this information. If you'd like to give it a try, you can download it here:
https://www.arduino.cc/en/software#experimental-software

PR #829, opened 2021-05-21
Update millis.adoc

Monzer-gom3a:


MalcolmBoura (commented):
This commit results in inconsistency between the code and the comment. 100ms in 0.1s, not 1s. It should be reverted.

PR #831, opened 2021-05-24
Change ''D" to "d" in <gives you the String "D">

brewmanz:
Perhaps the code has changed over time, but this documentation change is for the reference to reflect reality. 
Also consider changing the code to perhaps accept HEX_UPPER & HEX_LOWER.


per1234 (approved):
Thanks @brewmanz!

It's unfortunate that Arduino's String and Print classes are not consistent in this regard:
```cpp
Serial.println(String(13, HEX));  // d
char buf[2];
itoa(13, buf, 16);
Serial.println(buf);  // d
Serial.println(13, HEX);  // D
```

PR #860, opened 2021-12-24
Update compoundMultiplication.adoc

YClayton:
Change Example Code from:

x = 2;
x *= 2; // x now contains 4

----

Change it to:
x = 2;
x *= 3; // x now contains 6
----

Just makes it a little clearer, since 2+2=4 and 2*2=4. Too many twos.


per1234 (approved):
Thanks @YClayton!

PR #864, opened 2022-01-10
Strange results in floating point

Core-l:
6.0 / 3.0 != 2.0 is an incorrect example because floats can represent all three numbers perfectly, so the result is equal. 9.0 / 0.3 != 30.0 is a better example, it produces the correct result in double precision but not in single precision.


per1234 (approved):
Thanks @Core-l

PR #872, opened 2022-03-19
keyboardModifiers: split modifiers and special keys

### LINKED ISSUE #871: Keyboard: incorrect use of the term “modifier”
In the documentation of the Keyboard library, the page [Keyboard Modifiers][] provides a good definition of a “modifier key”:

> A modifier key is a special key on a computer keyboard that modifies the normal action of another key when the two are pressed in combination.

The page then lists “the definitions of the modifier keys”, namely a set of macros such as `KEY_LEFT_CTRL`, `KEY_RIGHT_SHIFT`, etc.

Although the first entries in this list are indeed modifier keys, most of the entries are not: `KEY_UP_ARROW`, `KEY_BACKSPACE`, `KEY_TAB`, `KEY_PAGE_UP`, `KEY_F1`, etc. This is both wrong and inconsistent with the definition provided in the same page.

[Keyboard Modifiers]: https://github.com/arduino/reference-en/blob/4ca291ef0a0c1564c4bd01f69b8a9c2c17001fc6/Language/Functions/USB/Keyboard/keyboardModifiers.adoc

edgar-bonet:
As stated in issue #871, the page [Keyboard Modifiers][] incorrectly uses the term “modifier” for many keys that are not real modifiers, such as `KEY_UP_ARROW`, `KEY_BACKSPACE`, etc. This pull request addresses the issue by splitting the table listing the “modifier keys” into two smaller tables:

- “Keyboard modifiers”, which now only lists the keys that are actually modifiers

- “Special keys”, which lists all the other macros: keys that are neither printable characters nor modifiers.

Fixes #871.

[Keyboard Modifiers]: https://github.com/arduino/reference-en/blob/4ca291ef0a0c1564c4bd01f69b8a9c2c17001fc6/Language/Functions/USB/Keyboard/keyboardModifiers.adoc

### DIFF
--- a/Language/Functions/USB/Keyboard/keyboardModifiers.adoc
+++ b/Language/Functions/USB/Keyboard/keyboardModifiers.adoc
@@ -1,11 +1,11 @@
 ---
-title: Keyboard Modifiers
+title: Keyboard Modifiers and Special Keys
 ---
 
 
 
 
-= Keyboard Modifiers
+= Keyboard modifiers and special keys
 
 
 // OVERVIEW SECTION STARTS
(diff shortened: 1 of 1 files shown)


per1234 (approved):
Thanks @edgar-bonet!

PR #873, opened 2022-03-26
update print.doc  -- rounding in case of floating point 

fujiwat:
In case of floating-point and specify the number of decimal places, Serial.print() supports rounding.  0 to 4 then rounddown, 5 to 9 then round up.
so the original text
Serial.print(1.23456, 4) gives "1.2345"
should be:
Serial.print(1.23456, 4) gives "1.2346"

I checked the result  on TINKERCAD and update line is correct.


per1234 (approved):
Thanks @fujiwat 

I see this is implemented in the standard Arduino core API here:

PR #874, opened 2022-04-13
Fix misspellings in headings

mattborja:
Headings fixed:
- "Syntax"
- "Parameters"


per1234 (approved):
Thanks @mattborja!

matthijskooijman:
I was going to leave an approving review, but seems I'm too late :-p

Thanks for the lightning quick action, both!

mattborja:
@matthijskooijman In fairness, 2 bytes of change doesn't take long to commit (or review for that matter) :P

matthijskooijman:
> @matthijskooijman In fairness, 2 bytes of change doesn't take too long to commit (or review for that matter) :P

Well, I think I might have had a couple of those bytes that took a *whole* lot longer in other projects, so no guarantees :-p

PR #888, opened 2022-06-20
Add missing function (endTransmission())

krytie75:
endTransmission function page exists but link is missing from the function list on the Wire.h page.


per1234 (approved):
Thanks so much for this important fix @krytie75!

PR #889, opened 2022-06-24
Update SPI.adoc

John-Karatka:
Added endTransaction() link to function section.


John-Karatka:
Signed CLA

John-Karatka (commented):
Signed CLA

per1234 (changes requested):
I think it will be best to place the link on the line after the link for the `beginTransaction` reference page, as the order was in the original SPI library reference before the loss of the link during the migration.

You can see how it used to look here on the Wayback Machine snapshot:

https://web.archive.org/web/20220120162953/https://www.arduino.cc/en/Reference/SPI

image

per1234 (approved):
Thanks so much for fixing this serious omission @John-Karatka!

PR #890, opened 2022-06-28
Modification of Returns section

Dreamy-Z3r0:
"The n’th character of the String" means the returned char is at index n+1 of the String, which is an incorrect description of the output.


per1234 (approved):
Thanks @Dreamy-Z3r0!

PR #892, opened 2022-07-06
Update PROGMEM.adoc

ka84:
fixed broken URLs


per1234 (approved):
Excellent work on finding the missing content @ka84!

In case it might be useful for future reference, I'll record the current URL of the page on the Arduino Playground:

https://playground.arduino.cc/Learning/Memory/

However, I think the page you linked to is best because it is more likely to be maintained than the archived Playground content.

PR #909, opened 2023-01-03
Create SimpleSub

Lobie23:
The  basics  to utilise  a  subroutine, a  new file Reference > Language > Structure > Sketch > (SimpleSub())
to help  those of us  that  need it, and   I obviously  was  not  the  only  person.
Needs  then to be  complimented with another file with  all  the optional passing  arguments.

### DIFF
--- a/Language/Structure/SimpleSub
+++ b/Language/Structure/SimpleSub
@@ -1,0 +1,28 @@
+SimpleSub()
+[Sketch]
+Description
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Thanks for your proposal @Lobie23. This is valuable information, but it is not in scope for the Arduino Language Reference so we will not be able to accept this addition.

You might consider publishing it on the Arduino Project Hub. The format of that platform even allows you to go into the subject matter more in depth than would be possible in the language reference.

PR #910, opened 2023-01-18
Update bit.adoc with allowed range of n

BloodyNewbie:
bit(n) with n > 31 do not return reasonable values. This should be mentioned

### DIFF
--- a/Language/Functions/Bits and Bytes/bit.adoc
+++ b/Language/Functions/Bits and Bytes/bit.adoc
@@ -28,7 +28,7 @@ Computes the value of the specified bit (bit 0 is 1, bit 1 is 2, bit 2 is 4, etc
 
 [float]
 === Parameters
-`n`: the bit whose value to compute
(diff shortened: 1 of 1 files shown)


Language/Functions/Bits and Bytes/bit.adoc
```diff
@@ -28,7 +28,7 @@ Computes the value of the specified bit (bit 0 is 1, bit 1 is 2, bit 2 is 4, etc
 
 [float]
 === Parameters
-`n`: the bit whose value to compute
+`n`: the bit whose value to compute. 0 <= n < 32 (as it is intended to reflect the bits up to 31 of a uint32)
```
karlsoderby, line 31:
```suggestion
`n`: the bit whose value to compute. Note that `n` needs to be between 0-31 (32 bit). 
```

Hi @BloodyNewbie , thanks for the submission. I modified it for easier readability. Let me know if you accept this, and I can merge it.

PR #920, opened 2023-08-11
Update abs.adoc

arjunchr:
just corrected a typing mistake.
i just tested example code and realized there is typing error in return part.
As its official documentation i think it should as accurate as possible.

### DIFF
--- a/Language/Functions/Math/abs.adoc
+++ b/Language/Functions/Math/abs.adoc
@@ -34,7 +34,7 @@ Calculates the absolute value of a number.
 [float]
 === Returns
 `x`: if x is greater than or equal to 0. +
-`-x`: if x is less than 0.
+`x`: if x is less than 0.
 
 --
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Hi @arjunchr. Thanks for taking the time to submit a pull request.

Please see https://github.com/arduino/Arduino/issues/3338#issuecomment-115157321 for an explanation of why the current documentation is correct.

## Related


- https://github.com/arduino/Arduino/issues/6214
- https://github.com/arduino/reference-en/pull/363
- https://github.com/arduino/reference-en/issues/597
- https://github.com/arduino/reference-en/pull/609
- https://github.com/arduino/reference-en/pull/841
- https://forum.arduino.cc/t/arduino-reference-abs-function/1026622

arjunchr:
ok, i understood can we add some extra line in documentation to clear the concept.

PR #950, opened 2023-12-06
Update Serial.adoc

Guswib:
The pins for uno R3 had been put in the wrong column. Now they are under serial.  Uno R4 was also updated to reflect that it looks like the Leonardo.


karlsoderby:
Hi there @Guswib , it turns out the information for all boards was inaccurately updated some days ago so really good that you spotted the error(s). I took the time to actually update the entire table, and in this process, I fixed what you suggested as well! You can see changes in this PR: https://github.com/arduino/reference-en/pull/953

Thanks for the submission 🚀

PR #957, opened 2023-12-23
Volatile example code variable name is now `changed`, not `state`

davidhbrown:
A previous version of this page had different example code. This updates the introduction to the example to use the current example's volatile variable name.


karlsoderby (approved):
LGTM, thanks for the submission @davidhbrown !

PR #968, opened 2024-02-13
Add word 'specify' to entries in switch...case parameter descriptions

AnturGyffrous:
The sentences in the parameter descriptions for the switch...case control structure documentation don't read well in my opinion, I think they need to use the word 'specify' e.g. "You can also use the `bool` data type when you **specify** just two switch cases."

I also think the code example should consistently use a space between the single line comment slashes and the comment text.


per1234 (approved):
Thanks @AnturGyffrous!

PR #971, opened 2024-02-15
Update highLow.adoc

drf5n:
Change the 0/1 order to match the HIGH/LOW & true/false order in the beginning of the sentence.


per1234 (approved):
Thanks @drf5n!

